PR #3, opened 2019-09-10
Add crates.io and docs.rs badges to README

Herschel:


kyren:
Oh hey, thanks for this!

PR #7, opened 2021-02-07
Add no std support

### LINKED ISSUE #1: no-std support?
Would this be a possible goal of this project?

madsmtm:
Fixes #1.

The extra tests in `gc-arena/src/context.rs` for the correct implementation of the rounding behaviour might be excessive?


kyren:
Yeah this all seems reasonable to me, thank you for this!

I wasn't even aware f64::round wasn't available on no_std, that's interesting (I don't do that much no_std work).

PR #11, opened 2021-02-12
Add #[track_caller] for better panic error messages

adrian17:
Before:
```
Error message: panicked at 'already borrowed: BorrowMutError', gc-arena-0.2.0/src/gc_cell.rs:62:29
```
After: it shows the line that actually tried calling read()/write().
```
Error message: panicked at 'already borrowed: BorrowMutError', core/src/avm1/activation.rs:1186:48
```
Should help a bit with pinpointing error sources for library users :)


madsmtm (approved):

kyren:
Yeah this makes sense to have.

PR #22, opened 2022-11-22
loosen struct def trait bounds

### LINKED ISSUE #21: Build Hangs Recursive Struct
I've been using this crate extensively for a `no-std` language vm. Part of my interface is as follows (minimized):

```rs
#[derive(Collect)]
#[collect(no_drop)]
pub enum Value<'gc> {
    List(Gc<'gc, Vec<Value<'gc>>>),
    Closure(Gc<'gc, Closure<'gc>>),
}

#[derive(Collect)]
#[collect(no_drop)]
pub struct Closure<'gc> {
    pub vars: Vec<Value<'gc>>,
}
```

This has worked perfectly fine for months. However, I recently made one little change to add a custom user type `S`, and suddenly `cargo build` and `cargo check` hang and eat up memory until it runs out and crashes. I've managed to track down the issue to this dependency and have a minimal example:

```rs
#[derive(Collect)]
#[collect(no_drop)]
pub enum Value<'gc, S> { // note the param S, which seems to cause the problem
    List(Gc<'gc, Vec<Value<'gc, S>>>),
    Closure(Gc<'gc, Closure<'gc, S>>),
}

#[derive(Collect)]
#[collect(no_drop)]
pub struct Closure<'gc, S> {
    pub vars: Vec<Value<'gc, S>>,
}
```

Worse, there is no error message whatsoever, so it was hard to track down. But I saw #4 and decided to try it. In my minimal example, this now gives a recursion depth error (better), but it doesn't solve my actual code base issues, which are more like the following again hanging example:

```rs
use gc_arena::{Collect, Gc};

#[derive(Collect)]
#[collect(no_drop)]
pub enum Value<'gc, S> {
    List(Gc<'gc, Vec<Self>>),
    Closure(Gc<'gc, Closure<'gc, S>>),
    Closure2(Gc<'gc, Closure2<'gc, S>>),
}

#[derive(Collect)]
#[collect(no_drop)]
pub struct Closure<'gc, S> {
    pub vars: Vec<Value<'gc, S>>,
}

#[derive(Collect)]
#[collect(no_drop)]
pub struct Closure2<'gc, S> {
    pub vars: Vec<Value<'gc, S>>,
}
```

Because adding a second non-self branch to the enum reintroduces the problem, I'm assuming this crate does some kind of breadth-first search for checking that a type's fields/variants satisfy `Collect` requirements (with a special case for `Self`), but which doesn't protect from cycles? Is there not a better way to check for cycles than a special case for (only) `Self`?

I need to be able to expose `Closure` instances to the library user, but I could possibly fix this trivial example by inlining the fields of `Closure` in `Value::Closure`, replacing `Value<'gc, S>` with `Self`, and reconstructing the actual `Closure` instance from the fields when requested by the user. However this would not solve my actual issues because the real (full) definition of `Closure` has several layers of indirect typing, all of which need to be exposed to users.

So my main questions are: why does the code without `S` not cause a problem, and is there some workaround to make this work?

dragazo:
Closes #21. It seems like the early requirement of `T: Collect` directly in the definition of several types (notably `Gc` and `GcCell`) was causing the rust compiler's trait bound satisfaction check to decay into infinite recursion for both direct and indirect recursive types. Simply removing this early requirement (on only the struct definitions) but keeping all the bounds on implementations solves the problem, and I believe doesn't impose any unsoundness since the struct does nothing on its own and all the implementators use (and need) `Collect` on their own.

### DIFF
--- a/src/gc-arena/src/context.rs
+++ b/src/gc-arena/src/context.rs
@@ -340,7 +340,7 @@ impl Context {
     unsafe fn write_barrier<T: Collect>(&self, ptr: NonNull<GcBox<T>>) {
         // During the propagating phase, if we are mutating a black object, we may add a white
         // object to it and invalidate the invariant that black objects may not point to white
-        // objects.  Turn black obejcts to gray to prevent this.
+        // objects.  Turn black objects to gray to prevent this.
         let gc_box = ptr.as_ref();
         if self.phase.get() == Phase::Propagate && gc_box.flags.color() == GcColor::Black {
             gc_box.flags.set_color(GcColor::Gray);
--- a/src/gc-arena/src/gc.rs
+++ b/src/gc-arena/src/gc.rs
@@ -13,7 +13,7 @@ use crate::types::{GcBox, Invariant};
 /// through "generativity" such `Gc` pointers may not escape the arena they were born in or be
 /// stored inside TLS.  This, combined with correct `Collect` implementations, means that `Gc`
 /// pointers will never be dangling and are always safe to access.
-pub struct Gc<'gc, T: 'gc + Collect> {
+pub struct Gc<'gc, T: 'gc> {
     pub(crate) ptr: NonNull<GcBox<T>>,
     _invariant: Invariant<'gc>,
 }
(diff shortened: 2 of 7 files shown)


dragazo:
Alternatively, I think it would fix the remaining issues if the derive macro could directly generate a compile error if it fails to derive rather than producing trait bounds. The whole issue seems to revolve around infinite loops when checking trait bounds, so avoiding those would probably fix the issue, but I'm not knowledgeable enough in proc macros to know if that's possible.

PR #24, opened 2022-11-23
Add more useful Debug, Pointer, and Display impls

redacted-moose:
This adds more useful Debug and Pointer impls for the Gc and GcCell types, and adds a Display impl for the Gc type.  These impls mirror the implementations for Rc and RefCell from the standard library.

Note: this does mean that you will no longer be able to derive Debug on types that contain a Gc ~or GcCell~ if they themselves contain types that don't implement Debug.  You still can manually implement Debug in this case however.

### DIFF
--- a/src/gc-arena/src/gc_cell.rs
+++ b/src/gc-arena/src/gc_cell.rs
@@ -1,5 +1,5 @@
 use core::cell::{BorrowError, BorrowMutError, Ref, RefCell, RefMut};
-use core::fmt::{self, Debug};
+use core::fmt::{self, Debug, Pointer};
 
 use crate::collect::Collect;
 use crate::context::{CollectionContext, MutationContext};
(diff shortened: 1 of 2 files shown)


src/gc-arena/src/gc_cell.rs
```diff
@@ -23,7 +23,30 @@ impl<'gc, T: Collect + 'gc> Clone for GcCell<'gc, T> {
...
+                impl Debug for BorrowedPlaceholder {
+                    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                        f.write_str("<borrowed>")
+                    }
+                }
+
+                fmt.debug_struct("GcCell")
+                    .field("value", &BorrowedPlaceholder)
+                    .finish()
+            }
```
redacted-moose, line 42: One alternative here is to display the pointer value instead of just printing '<borrowed>', which at least provides a bit more info.  Thoughts?

kyren: I am totally fine with "<borrowed>" personally, especially if that's what the standard library currently does for Rc

kyren:
> Note: this does mean that you will no longer be able to derive Debug on types that contain a Gc or GcCell if they themselves contain types that don't implement Debug. You still can manually implement Debug in this case however.

I think this is completely fine and normal.

src/gc-arena/src/gc_cell.rs
redacted-moose: Here's what the [standard library does for reference](https://doc.rust-lang.org/src/core/fmt/mod.rs.html#2618), so I'm inclined to just leave it as it is.

PR #27, opened 2022-11-24
Small documentation improvements

moulins:
Some slight improvements to the documentation of the `Arena` type:
- Move explanatory paragraphs from `make_arena` to `Arena`;
- Add a `Root<'lt, R>` type alias to simplify signatures;
- Remove useless `#[allow(unused)]` attributes.


kyren:
Oh, thank you for cleaning that up, I noticed that there were a lot of old unnecessary attributes from when the arena types were constructed entirely via macro.

PR #28, opened 2022-11-24
Make mutating Arena methods more flexible

moulins:
- Allow `Arena::mutate` to return references inside the object graph, as long as they don't mention the 'gc lifetime. This is safe because drops can only occurs while collecting, which requires exclusive access to the arena.
  - This has also the nice side-effect of reducing the number of lifetimes the user has to care about.
- Add `Arena::mutate_root`, which allows mutating the root set.


kyren:
All of this makes sense to me, and I agree it looks sound! Thank you very much!

PR #31, opened 2022-11-27
update CI to run tests with miri, sanitizers

kyren:


kyren:
I had to reduce the complexity of `repeated_allocation_deallocation` in order to get testing with miri to finish in CI in a reasonable amount of time

PR #34, opened 2022-11-30
Fix unsoundness in Hash{Map, Set}'s Collect implementation

moulins:
These implementations accept arbitrary `S: BuildHasher`, but don't trace the value at all.

This adds an extra `S: 'static` bound to prevent smuggling Gc pointers through the hasher. (All 'real' hashers are `'static`, so this is better than adding a `Collect` bound w.r.t. third-party hashers support)


kyren:
Ahhhh good catch!

PR #51, opened 2023-05-03
Get rid of a bad footgun with `Collect` impls on `&T`

kyren:
The existing `Collect` impls on reference types *are* sound, but they're a huge footgun because it's easy to accidentally call `<&T as Collect>::trace` instead of `T::trace`.

Just remove them and have an impl only for `&'static T`. Now, if you accidentally try to call `<&T as Collect>::trace`, you will at least get a compiler error instead of a silent footgun.


moulins:
LGTM; it would be best to pair this with a more powerful `StaticCollect` wrapper (e.g. implementing more traits), but this can wait for a future PR.

PR #52, opened 2023-05-05
Add a longer form version of the  macro to avoid lifetime name collision

kyren:


kyren:
This is a tiny change to make a common pattern that I have run into about 20 times today a lot easier.

May the github gods forgive me, I am merging without code review.

PR #53, opened 2023-05-06
Allow `Rootable` impls to have ?Sized Root types.

kyren:
Allows for dynamic rooting of unsized `Gc` pointers.

### DIFF
--- a/src/gc-arena/src/arena.rs
+++ b/src/gc-arena/src/arena.rs
@@ -66,7 +66,7 @@ impl ArenaParameters {
 /// `Rootable<'a>` for *any* possible `'a`. This is necessary so that the `Root` types can be
 /// branded by the unique, invariant lifetimes that makes an `Arena` sound.
 pub trait Rootable<'a> {
-    type Root: Collect + 'a;
+    type Root: ?Sized + Collect + 'a;
 }
 
 /// A convenience macro for quickly creating type that implements of `Rootable`.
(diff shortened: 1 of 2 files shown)


kyren:
We could add a `SizedRootable` trait as an "alias" for requiring that the associated `Root` type be sized, if we considered it obnoxious to add, but realistically I don't think that many people are going to be writing `Rootable` bounds?

kyren:
In https://github.com/kyren/deimos, this ended up being a bit more work than I thought due to the magic downcasting stuff there.

You would think it would just be adding `: Sized` bounds everywhere but for some reason I can't fully grasp, it made the compiler creak and groan a bit. I ended up having to drop a `Sized` bound on a `Collect` impl to make it work, and the error messages until I did that were extremely confusing and entirely about the `'gc` lifetime and I'm not sure why. I don't fully understand why it wasn't *exactly* the same, but it's possible I just missed a different `Sized` bound somewhere or messed something else up.

Adding an alias trait might actually be a good idea if we merge this.

kyren:
This is *significantly* less important now that #54 is merged, and in fact I don't know of a use case for it now?

kyren:
> You would think it would just be adding : Sized bounds everywhere but for some reason I can't fully grasp, it made the compiler creak and groan a bit. I ended up having to drop a Sized bound on a Collect impl to make it work, and the error messages until I did that were extremely confusing and entirely about the 'gc lifetime and I'm not sure why. I don't fully understand why it wasn't exactly the same, but it's possible I just missed a different Sized bound somewhere or messed something else up.

This was a false alarm, I just typoed a `'gc_` as `'gc`, I should probably not have picked those names.

This works fine, but still, I don't know if it's necessary or not.

kyren:
superseded by #102

PR #56, opened 2023-05-08
Expose inner `GcRefCell` as `CollectRefCell` and also add `CollectCell`

kyren:
These types are generally useful, even if they require unsafe to be useful.

It's nice to not have to manually make an unsafe `Collect` impl, rather than only having to remember to trigger a write barrier.


kyren:
If anyone has a better name than `CollectRefCell` and `CollectCell` I'm all ears.

moulins:
Instead of having a separate method for each unsafe operation on each cell type, I'd propose just providing a single method:
```rs
unsafe fn as_(ref_)cell(&self) -> &(Ref)Cell<T>;
```

This is slightly more verbose to use, but IMHO, it has two advantages:
- The user can call any method on the `Cell/RefCell`, not just those we explicitely support;
- Using a specific method name clarifies what "kind" of unsafety is happening.

---------

> If anyone has better names than `CollectRefCell` and `CollectCell` I'm all ears.

Some suggestions:
- `Lock/RefLock`: this is what I'm using in my [`gc-extra` experiment](https://github.com/moulins/ruffle/tree/gc-extra/gc_extra/src/lock);
- `Jail/RefJail`: cute variation on `Cell`;
- `FrozenCell/FrozenRefCell`: a little verbose, but maybe more explicit?

src/gc-arena/src/cell.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn get(&self) -> T {
+        self.cell.get()
+    }
+}
+
+unsafe impl<'gc, T: Collect + Copy + 'gc> Collect for CollectCell<T> {
+    fn trace(&self, cc: CollectionContext) {
+        self.cell.get().trace(cc);
+    }
+}
```
moulins, line 71: This is a very degenerate case, but this is incorrect if `T::trace` modifies the inner `T`; the potentially modified value should be written back to the `Cell`:
```rs
fn trace(&self, cc: CollectionContext) {
    // Use a guard that writes back the value in case the 'trace' call mutates it.
    let cell = &self.cell;
    let guard = scopeguard::guard(cell.get(), |copied| cell.set(copied));
    T::trace(&guard, cc);
}
```

PR #59, opened 2023-05-09
Expand upon `Lock` APIs

moulins:
A bunch of bells and whistles:
- move the lock types into their separate module, to not clutter the crate root;
- implement *all* the traits;
- add some extra methods (`into_inner`, `get_mut`, `take`);
- replace `Gc::write_(ref_)cell` by `Gc::unlock` and the `Unlock` trait;
- add type aliases for `Gc`+locks combinations.

I wanted to add `OnceLock` too, but I realized that `OnceCell` will only be stabilized in Rust 1.70, so I guess I'll have to wait a little still.

### DIFF
--- a/src/gc-arena/src/dynamic_roots.rs
+++ b/src/gc-arena/src/dynamic_roots.rs
@@ -4,7 +4,8 @@ use alloc::{
 };
 use core::mem;
 
-use crate::{Collect, Gc, MutationContext, RefLock, Root, Rootable};
+use crate::lock::RefLock;
+use crate::{Collect, Gc, MutationContext, Root, Rootable};
 
 // SAFETY: Allows us to conert `Gc<'gc>` pointers to `Gc<'static>` and back, and this is VERY
 // sketchy. We know it is safe because:
--- a/src/gc-arena/src/dynamic_roots.rs
+++ b/src/gc-arena/src/dynamic_roots.rs
@@ -53,7 +53,7 @@ impl<'gc> DynamicRootSet<'gc> {
         mc: MutationContext<'gc, '_>,
         root: Root<'gc, R>,
     ) -> DynamicRoot<R> {
-        let mut inner = self.0.write_ref_cell(mc).borrow_mut();
+        let mut inner = self.0.unlock(mc).borrow_mut();
 
         let handle = Rc::new(Handle {
             set_id: inner.set_id.clone(),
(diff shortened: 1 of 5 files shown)


kyren:
We discussed this a bit in Discord, just repeating myself here for future reference...

1) I *love* the name `unlock` to go with the `lock` types, makes it absolutely clear to me what they do.

2) The safety of `Lock::take` and `RefLock::take` have some real subtlety to them. The parametricity argument makes complete sense to me, but I pointed out an extreme corner case point that it does make potentially unsound code using `Default::default` have some... non-locality to it. One could imagine the unsafe internals of a module containing a `Default::default` that is doing some weird tricks to materialize a `MutationContext` from global state, and the existence of `RefLock::take` making it... "more unsafe" than it would otherwise be. This is an extremely fine point and it is very difficult to imagine the existence of `RefLock::take` making a sound API into an unsound one, especially since having such a `Default` impl is automatically unsound, but it might be *theoretically* possible. In any case, it's such a minor point that we agreed it's fine to just leave a slightly more detailed comment explaining the situation.

kyren:
This looks really good to me!

Do you think it would be worth it to provide a more convenient constructor on `Gc<Lock>` / `Gc<RefLock>` like the old `GcCell` had?

We have the `Gc[Ref]Lock` typedefs, but you still have to type e.g. `Gc::new(mc, RefLock::new(t))`. We could have `Gc::new_locked` or something to make it slightly more ergonomic? It might not even really be worth it, but just mentioning it as a suggestion.

Other than that and the extremely minor point before, LGTM!

moulins:
> We have the Gc[Ref]Lock typedefs, but you still have to type e.g. Gc::new(mc, RefLock::new(t)). We could have Gc::new_lock / Gc::new_ref_lock or something to make it slightly more ergonomic? It might not even really be worth it, but just mentioning it as a suggestion.

This already adds `impl<T> From<T> for [Ref]Lock<T>`, so you can use `Gc::new(mc, t.into())`, which (I think) is ergonomic enough.

------

P.S. Added a commit to expand on the safety comment of `[Ref]Lock::take`.

kyren:
> This already adds impl<T> From<T> for [Ref]Lock<T>, so you can use Gc::new(mc, t.into()), which (I think) is ergonomic enough.

Ah yeah, fair enough, that's probably plenty. LGTM!

PR #60, opened 2023-05-11
Inline trivial methods

moulins:
Adds `#[inline]` to trivial methods, so that the compile-time wrappers get properly optimized out without having to enable LTO.

### DIFF
--- a/src/gc-arena/src/context.rs
+++ b/src/gc-arena/src/context.rs
@@ -359,7 +359,15 @@ impl Context {
         let header = gc_box.header();
         if self.phase.get() == Phase::Propagate && header.color() == GcColor::Black {
             header.set_color(GcColor::Gray);
-            self.gray_again.borrow_mut().push(gc_box);
+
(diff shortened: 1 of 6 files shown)


src/gc-arena/src/context.rs
```diff
@@ -344,17 +351,27 @@ impl Context {
...
         if self.phase.get() == Phase::Propagate && header.color() == GcColor::Black {
             header.set_color(GcColor::Gray);
-            self.gray_again.borrow_mut().push(gc_box);
+
+            // Outline the actual enqueueing code (which is somewhat expensive and won't be
+            // executed often) to promote the inlining of the write barrier.
+            #[cold]
+            fn enqueue(this: &Context, gc_box: GcBox) {
+                this.gray_again.borrow_mut().push(gc_box);
+            }
```
kyren, line 368: It's probably very minor, but should `header.set_color(GcColor::Gray)` also be in `enqueue`?

kyren:
Went ahead and merged it, it's fine as is really.

PR #63, opened 2023-05-16
Workaround for a rustc bug that can currently lead to UB

kyren:
Credit goes to me for stumbling across it and 100% to @moulins for actually finding the fix.

I do not have a link to an open rust issue for the specific bug, I can try and find a link later if @moulins doesn't already know which one it is.

### DIFF
--- a/src/gc-arena/src/collect_impl.rs
+++ b/src/gc-arena/src/collect_impl.rs
@@ -75,7 +75,33 @@ static_collect!(std::ffi::OsStr);
 #[cfg(feature = "std")]
 static_collect!(std::ffi::OsString);
 
-unsafe impl<T: ?Sized> Collect for &'static T {
+/// SAFETY: We know that a `&'static` reference cannot possibly point to `'gc` data, so it is safe
+/// to keep in a rooted objet and we do not have to trace through it.
+///
+/// HOWEVER, There is an extra bound here that seems superfluous. If we have a `&'static T`, why do
+/// we require `T: 'static`, shouldn't this be implied, otherwise a `&'static T` would not be well-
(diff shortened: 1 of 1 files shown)


moulins:
As extra evidence that this is an actual `rustc` bug and not an issue with `gc-arena`'s abstractions, here's a piece of code using a similar trait setup and `Any` to demonstrate UB with no unsafe code ([playground link](https://play.rust-lang.org/?version=beta&mode=release&edition=2021&gist=5bcc2bf8b258046330c84525e2eb7663)):
```rust
use std::any::Any;
use std::sync::OnceLock;

trait Visit {
    fn visit(&self);
}

trait Visitable<'a> {
    type Visit: Visit + 'a;
}

static HOLDER: OnceLock<&'static String> = OnceLock::new();

impl<T> Visit for &'static T {
    fn visit(&self) {
        let any = (*self) as &dyn Any;
        if let Some(i) = any.downcast_ref::<&'static String>() {
            HOLDER.get_or_init(|| i);
        }
    }
}

fn call_and_visit<V, F>(value: String, f: F)
where
    V: for<'a> Visitable<'a> + ?Sized,
    F: for<'a> Fn(&'a String) -> <V as Visitable<'a>>::Visit,
{
    f(&value).visit();
}

type MyVisitable = dyn for<'a> Visitable<'a, Visit = &'static &'a String>;

fn main() {
    call_and_visit::<MyVisitable, _>("hello".into(), |i| {
        Box::leak(Box::new(i))
    });

    // oops, use-after-free
    dbg!(HOLDER.get());
}
```

I believe this is an instance of https://github.com/rust-lang/rust/issues/84533, as it also involves "broken" `dyn Trait` types allowing the bypass of well-formed checks.

PR #70, opened 2023-06-01
Fix `Rootable!` not compiling if not directly imported

moulins:
The missing `$crate` prefix was preventing `gc_arena::Rootable![Foo<'gc>]` from compiling.

-------

Relatedly, and following our discussion on Discord a few weeks ago, should I change the name of the "default lifetime" of the short form to `'__`, or `'_`? (the latter would be most natural, but would require a proc-macro) This would make the short form usable in most scenarios, when a `'gc` lifetime is already in scope.

### DIFF
--- a/src/gc-arena/src/arena.rs
+++ b/src/gc-arena/src/arena.rs
@@ -124,7 +124,7 @@ macro_rules! Rootable {
         $crate::__DynRootable::<dyn for<$gc> $crate::Rootable<$gc, Root = $root>>
     };
     ($root:ty) => {
-        Rootable!['gc => $root]
+        $crate::Rootable!['gc => $root]
     };
 }
 


kyren:
Yeah forgetting the `$crate`  prefix is entirely my fault, good catch.

I definitely think the `'gc` default for the short version of the macro is wrong because I find myself using the long form of the macro almost *exclusively*. The only thing that's giving me pause about using `'_` via proc-macro is that that lifetime already has another meaning, but I can't for the life of me fathom a case where the actual placeholder lifetime would be useful there, AND the concept of the placeholder lifetime makes sense in this context already. So, I guess by that logic, the placeholder lifetime makes the MOST sense? If it's pretty easy to write that proc-macro I guess that'd be the ideal way?

PR #71, opened 2023-06-02
Use elided lifetimes instead of `'gc` for the short form of `Rootable!`

moulins:
This allows using the short form even when a `'gc` lifetime is already in scope.

For ease of future maintenance, the lifetime substitution is done by a procedural macro instead of a recursive tt-muncher.


kyren:
LGTM! I am soooo tired of typing the full `Rootable!` incantation, and this makes it a lot clearer to me and less visually noisy too.

PR #72, opened 2023-06-04
`DynamicRootSet` additions

moulins:
- Adds `DynamicRootSet::{contains, try_fetch}`;
- Documents `DynamicRoot{, Set}`;
- Use `unlock!` to remove `RefCell` runtime checks during fetching.

### DIFF
--- a/src/gc-arena/src/dynamic_roots.rs
+++ b/src/gc-arena/src/dynamic_roots.rs
@@ -8,6 +8,10 @@ use crate::barrier::unlock;
 use crate::lock::RefLock;
 use crate::{Collect, Gc, Mutation, Root, Rootable};
 
+/// A way of registering GC roots dynamically.
+///
+/// Use this type as (a part of) an [`Arena`](crate::Arena) root to enable dynamic rooting of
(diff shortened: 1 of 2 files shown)


kyren:
LGTM, I almost asked why we can't use the automatic `Collect` impl on `DynamicRootSet` before I remembered that the `Collect` derive doesn't work within the gc-arena crate itself 🤦‍♀️. It's probably not worth fixing that since the implementations are so minimal, but it is kind of annoying.

Veykril:
Fwiw, it would be pretty easy to make the derive macro work in gc-arena itself by just adding an `extern crate self as gc_arena;` alias to the crate root of gc-arena.

Though its questionable whether you should use it if you don't have to since that would make gc-arena itself depend on syn.

kyren:
Yeah I know that would make it work and I thought about that, it's just always a pretty minor thing so it never feels worth it.

moulins:
Using the derive macro would only allow getting rid of the trivial `impl Collect for DynamicRootSet`, so it's not even that useful IMHO.

PR #75, opened 2023-06-29
Implement Eq, PartialEq, Ord, PartialOrd and Hash for Gc<T>

Phantomical:
This implements the above traits for `Gc<T>` when `T` implements them. I have tried to model the trait impls after how std implements them for `Rc<T>`. There are probably more traits that could be added here. I didn't look too hard at the impls list for `Rc` and just implemented the few that seemed the most useful.

My use case for this is that I would like to use `Gc<Path>`  as keys in a hashmap.


kyren:
LGTM, thank you!

PR #76, opened 2023-07-11
Tracked external allocations

kyren:
Extremely simplistic design, tracks external allocations as a bytecount, increasing the total external allocations speeds up collection, decreasing it slows down collection, both occur the same as if memory was allocated in the "normal" fashion.

External memory owned per gcbox is not tracked, instead, for the purposes of pacing calls to `trace`, each gcbox is assumed to own an equal fraction of the total external bytecount.

### DIFF
--- a/src/gc-arena/src/arena.rs
+++ b/src/gc-arena/src/arena.rs
@@ -2,7 +2,7 @@ use alloc::boxed::Box;
 use core::{f64, marker::PhantomData, usize};
 
 use crate::{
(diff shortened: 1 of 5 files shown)


kyren:
Addressed a bunch of stuff over discord, the math is very difficult to review but it at least APPEARS not to be *broken*. This is going to live in master for a while while we build tooling to do better testing and gather more data and stop working on this kind of thing so blind.

PR #77, opened 2023-07-13
Collect for allocator-api2 and hashbrown, also Metrics allocator

kyren:
Implements `Collect` for HashMap and HashSet from hashbrown, also Vec and Box from allocator-api2. Also implements a custom `Allocator` trait wrapper that records allocations and deallocations from gc-arena's `Metrics` handle, and is branded with the branding lifetime so it cannot be mixed up with other arena types.

The implementation for hashbrown is slightly subtle because it is different depending on whether or not the "allocator-api2" feature is also enabled. Circle-CI now also checks all combinations of the features "allocator-api2" and "hashbrown" to make sure they all compile.

This also subsumes #73 as it it's hard to disentangle them

### DIFF
--- a/src/gc-arena/src/allocator_api.rs
+++ b/src/gc-arena/src/allocator_api.rs
@@ -104,12 +104,7 @@ unsafe impl Collect for Global {
     }
 }
 
-unsafe impl<T: Collect, A: Collect + Allocator> Collect for boxed::Box<T, A> {
-    #[inline]
-    fn needs_trace() -> bool {
-        T::needs_trace() || A::needs_trace()
-    }
-
+unsafe impl<T: Collect + ?Sized, A: Collect + Allocator> Collect for boxed::Box<T, A> {
     #[inline]
     fn trace(&self, cc: &Collection) {
         boxed::Box::allocator(self).trace(cc);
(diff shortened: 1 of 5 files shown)


kyren:
Marking as a draft until I at least check it in piccolo..

src/gc-arena/src/allocator_api.rs
```diff
@@ -104,12 +104,7 @@ unsafe impl Collect for Global {
     }
 }
 
-unsafe impl<T: Collect, A: Collect + Allocator> Collect for boxed::Box<T, A> {
-    #[inline]
-    fn needs_trace() -> bool {
-        T::needs_trace() || A::needs_trace()
-    }
-
+unsafe impl<T: Collect + ?Sized, A: Collect + Allocator> Collect for boxed::Box<T, A> {
```
moulins, line 107: It's a little unfortunate that this looses the `needs_trace` optimization, but fixing this would require either a non-object-safe `Collect` trait, or `Collect::needs_trace(&self)`, so oh well...

kyren: Yeah.. it is unfortunate, but it's the same for the normal `Box` impl too.

kyren:
I have working memory tracking in piccolo with this and also #79, it works so I'm unmarking it as a draft.

It can be merged as is, but this is a little in the way atm of effectively using it https://github.com/zakarumych/allocator-api2/pull/5

PR #79, opened 2023-07-13
Count all allocations used by `DynamicRootSet` as external bytes.

kyren:
This includes bytes allocated by individual root handles.

When the API becomes "rootless" (which imo should happen before the next release), this will make gc-arena theoretically count the *total* memory used by the arena including all roots.

### DIFF
--- a/src/gc-arena/src/dynamic_roots.rs
+++ b/src/gc-arena/src/dynamic_roots.rs
@@ -4,8 +4,11 @@ use alloc::{
 };
 use core::{fmt, mem};
 
-use crate::barrier::unlock;
 use crate::lock::RefLock;
+use crate::{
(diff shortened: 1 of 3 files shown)


kyren:
I went back and forth on this, I don't think incrementing the external byte count is useful from the perspective of *collecting* anything, because we know that since those are *roots* we cannot possibly collect them by running the collector more. However, since they are roots they exist "inside" the arena and are morally allocated by it, and used to actually be allocated by it when they used `Gc` pointers instead of `Rc`.

I think from the perspective of "keeping memory usage in check" it MIGHT make sense to keep track of root sizes as allocations?

I really can't make up my mind on this, but I opened this PR to discuss it.

PR #80, opened 2023-08-25
Add methods to project `Write<Option<_>>` and `Write<Result<_, _>>`

moulins:
Some dumb `Write` projection methods I found the need for while hacking on Ruffle; ideally there would be a generic safe destructuring macro supporting enums, but this is much simpler to implement ^^


kyren:
Yeah, it would be nice to have a write macro for enums but even in the presence of such a macro, these methods are hardly harmful, lgtm!

PR #81, opened 2023-08-27
Fix building on platforms without Arc

dragazo:
`alloc::sync::Arc` only exists on platforms with atomic pointer-size operations. Our inclusion of `Collect` for `Arc<T>` was unconditional, so this was causing compile errors on platforms that don't have such atomic operations. I just added a cfg so we can keep it on all platforms that allow it.


kyren:
Makes sense to me!

PR #91, opened 2023-12-20
Add GcWeak::trace_strong

kyren:
Allows someone to more easily decide at runtime whether a GcWeak should be treated as a strong pointer or a weak pointer during tracing.

This is not *strictly* required for `piccolo` to implement weak tables but it makes it a lot less obnoxious.

Includes a couple of drive by changes adding extra debug asserts, and a *very small* perf increase for GcWeak::resurrect.

### DIFF
--- a/src/gc-arena/src/context.rs
+++ b/src/gc-arena/src/context.rs
@@ -290,6 +290,7 @@ impl Context {
                         }
 
                         let guard = DropGuard { cx: self, gc_box };
+                        debug_assert!(gc_box.header().is_live());
                         unsafe { gc_box.trace_value(self.collection_context()) }
                         gc_box.header().set_color(GcColor::Black);
                         mem::forget(guard);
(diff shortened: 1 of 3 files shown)


kyren:
After discussing this on Discord, we felt that it was too much of a footgun, and allowed you to do some confusing (if memory safe) things without understanding the consequences. Since using this method always requires implementing `Collect::trace`, which is unsafe, you can accomplish the same thing by carefully using `GcWeak::from_ptr` / `Gc::from_ptr` to swap between logically holding a `GcWeak` or a `Gc`. It is always safe to go from `Gc` to `GcWeak`, but going the other way requires a successful upgrade *and* a write barrier on the containing pointer to ensure that the object is re-traced, since it is (morally speaking) a mutation causing new values to be owned. Requiring the user to do this does not require (much) more unsafety, and is conceptually more consistent.

PR #94, opened 2024-02-28
add Collect impl for `std::collections::LinkedList`

Jengamon:
Gotta collect them all.

Might not be incredibly necessary, tho...


kyren:
I mean, this makes sense to me and is harmless to add

PR #96, opened 2024-04-22
Add a way to deterministically transition to `CollectionPhase::Collecting`

kyren:
This is necessary when trying to use finalization, because you want to be completely *out* of the marking phase once you have decided finalization is finished for this cycle.

### DIFF
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -2,7 +2,7 @@ use alloc::boxed::Box;
 use core::{f64, marker::PhantomData};
 
 use crate::{
-    context::{Context, Finalization, Mutation, Phase},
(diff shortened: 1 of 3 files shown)


kyren:
Changed `MarkedArena::collect` to `MarkedArena::start_collecting`

Added a drive by change from `self.metrics.allocation_debt() <= target_debt` to `!(self.metrics.allocation_debt() > target_debt)` out of NaN paranoia

Added an `EarlyStop` enum for `Context::do_collection`, rather than relying on another custom method.

Changed `MarkedArena` to own a `&mut Arena` instead of `&Arena`. These are equivalent since Rust can't downgrade mutable borrows, but it expresses the intent better, and `MarkedArena` now actually calls `Context::do_collection`.

PR #100, opened 2024-06-26
Let Gc::cast and Gc::from_ptr work for ?Sized types

kyren:


kyren:
ptr::byte_offset requires rust 1.75, so this is a MSRV change.

PR #102, opened 2024-06-27
Remove `Collect + Sized` on `Rootable::Root` type.

kyren:
Since we are using `Gc` pointers for `DynamicRootSet` again, we need to relax the way the `Rootable` macro works so that it can work for `Gc<dyn Trait>` and `Gc` pointers which have been cast to other types.

This is a breaking change (ofc), but hopefully it should not cause too much churn?

We could add a trait alias for `Sized + Collect` Root types to make things much easier, but I think the only way to have an actually *helpful* trait alias is to use the very recently stable [associated type bounds](https://github.com/rust-lang/rust/pull/122055/).

I think it's probably just not a big deal unless you do deep tricks with `Rootable` (like `piccolo`'s `Any`).

This also removes the `'static` bound on `Rootable` because absolutely nothing was using it (`piccolo` does, but it can add it itself).

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -394,7 +394,7 @@ fn test_dynamic_roots() {
     let rc_a = Rc::new(12);
     let rc_b = Rc::new("hello".to_owned());
 
-    let mut arena: Arena<Rootable![DynamicRootSet<'_>]> = Arena::new(|mc| DynamicRootSet::new(mc));
+    let mut arena = Arena::<Rootable![DynamicRootSet<'_>]>::new(|mc| DynamicRootSet::new(mc));
 
     let root_a = arena
         .mutate(|mc, root_set| root_set.stash::<Rootable![Rc<i32>]>(mc, Gc::new(mc, rc_a.clone())));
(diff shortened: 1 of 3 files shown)


tests/tests.rs
```diff
@@ -456,9 +456,8 @@ fn test_dynamic_roots() {
 #[test]
 #[should_panic]
 fn test_dynamic_bad_set() {
-    let arena1: Arena<Rootable![DynamicRootSet<'_>]> = Arena::new(|mc| DynamicRootSet::new(mc));
-
-    let arena2: Arena<Rootable![DynamicRootSet<'_>]> = Arena::new(|mc| DynamicRootSet::new(mc));
```
kyren, line 461: One thing I noticed here is that `let arena: Arena<Rootable![MyRoot<'_>]> = Arena::new(|mc| ...);` no longer works, but `let arena = Arena::<Rootable![MyRoot<'_>]>::new(|mc| ...);` does, and I have no idea why.

It's a very minor paper cut, but it'd be nice not to break it.

kyren:
Fixing `piccolo` after this was not a big deal in the slightest, I don't know where my head was when I was testing #53.

PR #104, opened 2024-06-29
Simplest possible forward write barriers.

kyren:
This has some additional drive-by cleanup to make things more
consistent. It adds two methods to `Mutation`, one for backwards write
barriers (previously the only kind) and one for forward write barriers.

Both backwards and forward write barriers are changed to take one or
optionally two pointers: the parent and the adopted child. If both
pointers are given, then the barriers check if the parent is black and
the child is white, and if so, invoke the barrier. The backwards write
barrier will turn the parent gray, and the forwards write barrier will
turn the child gray.

If the backwards write barrier is only given the parent pointer, then it
unconditionally turns it gray if it was black, allowing it to adopt any
pointers at all. If the forwards write barrier is only given the child
pointer, then it unconditionally turns it gray if it was white, allowing
it to be adopted by any possible pointer.

This is based on #103 because it requires `Gc::erase` to be less obnoxious to use.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -467,6 +467,10 @@ impl Context {
         // During the propagating phase, if we are mutating a black object, we may add a white
         // object to it and invalidate the invariant that black objects may not point to white
         // objects. Turn the black parent object gray to prevent this.
+        //
+        // NOTE: This also adds the pointer to the gray_again queue even if `header.needs_trace()`
+        // is false, but this is not harmful (just wasteful). There's no reason to call a barrier on
+        // a pointer that can't adopt other pointers, so we skip the check.
         if self.phase.get() == Phase::Mark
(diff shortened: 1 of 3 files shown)


kyren:
There is a separate commit for forward write barriers on arbitrary `Collect` impls,  2293486f04042df5a2553eed8d5180abb84fcab9

It works fine and is sound (and is what @moulins and I had discussed originally as a MVP for forward barriers), but the API is a bit footgunny.

I left the simpler version of `Mutation::forward_barrier` there for simplicity when only dealing with `Gc` pointers (besides not having the slight "wrong reference" risk, `Gc::trace` does not panic, so there is also no panic risk).

If for some reason we decide we don't like the the "trace immediately" version of forward barriers, we can drop this commit.

PR #105, opened 2024-07-09
API compatibility breaks

kyren:
* Renames `CollectionPhase::Collecting` -> `CollectionPhase::Sweeping`
* Renames `StaticCollect` -> `Static`

### DIFF
--- a/examples/linked_list.rs
+++ b/examples/linked_list.rs
@@ -1,4 +1,4 @@
-use gc_arena::{lock::RefLock, Arena, Collect, Gc, Mutation, Rootable};
+use gc_arena::{Arena, Collect, Gc, Mutation, RefLock, Rootable};
 
 // We define a node of a doubly-linked list data structure.
 //
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -380,10 +380,14 @@ where
     }
 }
 
-/// Create a temporary arena without a root object and perform the given operation on it. No garbage
-/// collection will be done until the very end of the call, at which point all allocations will
-/// be collected.
-pub fn rootless_arena<F, R>(f: F) -> R
+/// Create a temporary arena without a root object and perform the given operation on it.
+///
+/// No garbage collection will be done until the very end of the call, at which point all
+/// allocations will be collected.
+///
+/// This is a convenience function that makes it a little easier to quickly test code that uses
+/// `gc-arena`, it is not very useful on its own.
+pub fn rootless_mutate<F, R>(f: F) -> R
 where
     F: for<'gc> FnOnce(&'gc Mutation<'gc>) -> R,
 {
--- a/src/arena.rs
+++ b/src/arena.rs
@@ -54,13 +54,13 @@ impl<'a, T: ?Sized + Rootable<'a>> Rootable<'a> for __DynRootable<T> {
 /// parameters, though in complex cases it may be better to implement `Rootable` directly.
 ///
 /// ```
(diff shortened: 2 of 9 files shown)


kyren:
The next version of `gc-arena` will be a 0.x version bump because the way `DynamicRootSet` works has changed.

Let's think about some minor API breaks we want to make before the next release!

The big other one I can think of is not re-exporting so much in the crate root.

@moulins thoughts?

moulins:
I would consider axing the `unsafe_empty_collect!` macro; it's not used anywhere inside the crate, and I struggle to find a situation when it would be the correct thing to use over `static_collect!`.

In regard to reducing root-level exports, maybe this should be paired with a dedicated `gc_arena::prelude` that can be glob-imported. I'm not entirely sure what items should be in this prelude, though.

Finally, the `Collect` trait as it currently exists precludes future optimizations:
- `Collect::needs_trace` is a method, which means we can't use the information statically;
- `Collect::trace` takes a `&Collection`, which means we need to lock/unlock a `RefCell` every time we want to push a `GcBox` onto the gray queues.

I propose the following API:
```rs
#[derive(Copy, Clone)]
pub struct CollectFlags {
  needs_trace: bool,
  // could add other flags here, e.g. needs_drop?
};

impl CollectFlags {
  pub const DEFAULT: Self = CollectFlags { needs_trace: true };
  pub const EMPTY: Self = CollectFlags { needs_trace: false };
  
  // used by the `Collect` derive-macro to combine field flags
  pub const fn combine(flags: &[Self]) -> Self { ... }
}

pub struct Collection<'a> {
  // could become `&'a mut Context` without breaking the `Collect::trace` signature
  cx: &'a Context,
  // not really necessary, but possibly useful in the future?
  _inv: Invariant<'a>,
}

impl Collection<'_> {
  pub fn trace<T: Collect + ?Sized>(v: &T) { ... }
  
  // equivalent to `Self::trace(&gc)`, but avoids the `Deref` footgun
  pub fn trace_gc<T: ?Sized>(gc: Gc<'_, T>) { ... }
}

pub trait Collect {
  const FLAGS: CollectFlags;
  
  fn trace(&self, cc: Collection<'_>);
}
```

It does make `Collect` not object-safe, but I don't believe `dyn Collect` is ever useful, so I'm quite OK with that.

kyren:
Yeah these are great suggestions. Full agreement about `unsafe_empty_collect`.

I think making the `Collection` object able to hold a `&mut Context` is also really good, but I think I'd like to go ahead and *make* that change at the same time as changing the API, which I think I can do in a separate PR soon.

As far as re-exporting less in the crate root, I don't know what would go in such a prelude module but it's probably not the current set. The issue I have with the current set of exports is that it's just kind of random, you *almost always* need a `Lock` or `RefLock`, and you *almost never* need `rootless_arena`, but the first two are not in the crate root and the second one is for some reason.

Mostly I just want this to change. We don't export *that* many things total, so my instinct is to put whatever would go in the prelude in the crate root, but change what's there now to make better sense, but honestly I'm fine with anything. Personally I actually don't *ever* use glob imports, so I'd ignore the existence of a prelude and just import everything manually.

kyren:
I made an attempt at a slightly better crate organization in 8b57ff3493ba6eb39f12ce92ecdff8c5d7a871ad.

I didn't make a prelude module or make any drastic changes at all really, just cleaned up the crate root *slightly*. I'm obviously biased, but I left everything in the crate root that's something that I personally reach for all the time and nothing else. Well, that's not entirely true because I also use the `Root` typedef a *lot* but I think everywhere you need that you're doing something pretty specialized.

PR #110, opened 2024-08-28
Remove pub from 'inner' field in write barrier

Nilando:
I think that because the inner field is pub, external code could potentially modify it which would be unsafe. By making inner private it also removes the need for `struct Write` to be non_exhaustive.

However, this does introduce some weirdness in needing an 'explicit' deref function to be used in the `field!`macro as well as a new way of asserting that the $value is of type `Write`.

### DIFF
--- a/src/barrier.rs
+++ b/src/barrier.rs
@@ -168,9 +168,9 @@ macro_rules! __field {
         // - similarly, the `__from_ref_and_ptr` method takes both a reference (for the lifetime)
         //   and a pointer, causing a compilation failure if the first argument was coerced.
         {
-            let _: &$crate::barrier::Write<_> = $value;
+            let value: &$crate::barrier::Write<_> = $value;
 
-            match $value.get_inner() {
+            match value.get_inner() {
                 $type { ref $field, .. } => unsafe {
                     $crate::barrier::Write::__from_ref_and_ptr($field, $field as *const _)
                 },
--- a/src/barrier.rs
+++ b/src/barrier.rs
@@ -171,8 +171,9 @@ macro_rules! __field {
             let _: &$crate::barrier::Write<_> = $value;
 
             match $value.get_inner() {
-                $type { ref $field, .. }
-                    => unsafe { $crate::barrier::Write::__from_ref_and_ptr($field, $field as *const _) },
+                $type { ref $field, .. } => unsafe {
+                    $crate::barrier::Write::__from_ref_and_ptr($field, $field as *const _)
+                },
             }
         }
     };
--- a/src/barrier.rs
+++ b/src/barrier.rs
@@ -14,10 +14,7 @@ use crate::Gc;
 #[non_exhaustive]
 #[repr(transparent)]
 pub struct Write<T: ?Sized> {
-    // Public so that the `field!` macro can pattern-match on it; the `non_exhaustive` attribute
-    // prevents 3rd-party code from instanciating the struct directly.
-    #[doc(hidden)]
-    pub __inner: T,
+    inner: T,
 }
 
 impl<T: ?Sized> Deref for Write<T> {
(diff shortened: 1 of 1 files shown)


Nilando:
Not really sure what caused the failing tests output to change... I'm sure it has to do with changing the `field!` macro, but I think the output is still acceptable.

moulins:
> I think that because the inner field is pub, external code could potentially modify it which would be unsafe.
Third-party code accessing `__inner` (or constructingg a `Write<_>` by value) isn't actually unsafe, only undesirable and mostly useless. As such, I don't think this change is necessary.

Additionally, your version of the `field!` macro evaluates its argument twice, which definitely shouldn't happen.

nilandasd:
> > I think that because the inner field is pub, external code could potentially modify it which would be unsafe.
> > Third-party code accessing `__inner` (or constructingg a `Write<_>` by value) isn't actually unsafe, only undesirable and mostly useless. As such, I don't think this change is necessary.
> 
> Additionally, your version of the `field!` macro evaluates its argument twice, which definitely shouldn't happen.

Oh I didn't realize that! I think a fix is to save it to a variable? like this?
```
{
            let value = $value;
            let _: &$crate::barrier::Write<_> = value;

            match value.get_inner() {
                $type { ref $field, .. } => unsafe {
                    $crate::barrier::Write::__from_ref_and_ptr($field, $field as *const _)
                },
            }
        }
```

moulins:
Oh, my first objection (and most important one!) was eaten by the quote formatting, let me copy it again:

Third-party code accessing  `__inner` (or constructing a `Write<_>` by value) isn't actually unsafe, only undesirable and mostly useless. As such, I don't think this change is necessary.

kyren:
> Third-party code accessing __inner (or constructing a Write<_> by value) isn't actually unsafe, only undesirable and mostly useless. As such, I don't think this change is necessary.

Full agreement, accessing a double underscore variable which is `#[doc(hidden)]` is not part of the API at all, so it's not possible *by definition* for it to be an unsafe interface (since it is not even an interface).

The goal of a safe interface is not to protect against industrial espionage, and the compiler is not supposed to be a security boundary.

kyren:
> Third-party code accessing __inner (or constructing a Write<_> by value) isn't actually unsafe, only undesirable and mostly useless. As such, I don't think this change is necessary.

I actually didn't fully appreciate moulins' point either, which is that EVEN IF you consider the `__inner` field to be a public interface, it's STILL (arguably) not unsound.

If you simply construct a `Write` *value*, then this gives you no extra powers that you didn't already have. If you access the `__inner` field of an existing `Write` then same deal, you could already do that via `Deref` / `DerefMut`.

The only way to get a dangerous `&Write<T>` is through `mem::transmute`-ing a `&T` with unsafe interior mutability that *exists inside a `Gc`*.

I suppose if the definition of `Write` was...

```
/// A transparent wrapper of anything at all with no guarantees whatsoever.
///
/// You can freely call `mem::transmute` to convert a `&T` to a `&Write<T>` if you want.
#[repr(transparent)]
pub struct Write<T: ?Sized> {
    pub inner: T,
}
```

Then *that* would be unsound..... documentation? But that's not what the documentation says, and the inner field is specifically documented to not be part of the public API at all.

So yeah, this makes for a stronger case against needing to change anything.

kyren:
Even if this isn't merged, I appreciate the PR and I *definitely* appreciate any and all effort looking at `gc-arena`'s API and trying to find unsoundness, so thank you.

I don't *think* this is an example of unsoundness, but still... many eyes are a good thing!

I'll leave this open in case anyone can figure out a way that accessing `__inner` without additional unsafe code can lead to potential UB, but *even then* I still think that a `#[doc(hidden)]` public field should not really *count*.

Nilando:
> Oh, my first objection (and most important one!) was eaten by the quote formatting, let me copy it again:
> 
> Third-party code accessing `__inner` (or constructing a `Write<_>` by value) isn't actually unsafe, only undesirable and mostly useless. As such, I don't think this change is necessary.

My mistake, I do believe you're right! For some reason I was thinking the write barrier worked a different way

PR #116, opened 2024-12-03
Further `Collect` trait improvements

kyren:
Builds off of #107 to further improve the `Collect` trait.

There are three main API changes:

* Makes the `Collect::trace` method take a generic parameter that must implement the new trait `Trace`. This will be useful in the future to generically scan any value type for the properties of their held pointers.
* Adds a `'gc` lifetime parameter to `Collect` and `Trace`, allowing for more safety especially when implementing `Trace` which is a safe trait.
* Changes the `#[derive(Collect)]` proc macro to handle the required `'gc` lifetime on `Collect` and `Trace`. Almost always this will work without any further changes, but in the extremely rare and almost completely useless case where a type has multiple lifetime parameters, one must now be selected with `#[collect(gc_lifetime = 'gc)]`.

There are also some additional related internal changes
* Internally, don't use `RefCell<Vec<GcBox>>` for the gray queues. Instead, just unsafety via a new internal `Queue` type. This avoids the need for locking in all cases.
* Because of the previous change, we can get rid of `MarkState` entirely and implement `Trace` on `Context` directly.

This also changes some parts of the design of #107 and there are some open questions.

* `Collect::trace` now accepts `&mut (impl Trace<'gc> + ?Sized)` instead of `impl Trace<'gc>` which would more closely match what used to be `Collection<'_>`. This is for a few reasons:

  1) Requiring the implementer of `Trace` to always provide a special reborrow trace method that takes `T: Collect` is slightly obnoxious, and would require that every implementation always include the `if T::NEEDS_TRACE { ... }` check itself.
  2) We probably want to leave open the possibility of passing a dyn trait here, in case we need to one day need to allow tracing the *contents* of a `Gc` via vtable (such as with a cast `Gc` pointer).
  3) Every method on `Trace` accepts `&mut self`, so accepting a `&mut impl Trace<'gc>` allows for avoiding a double indirection on `Trace` methods.

  However, this does not *enforce* that implementers of `Collect` must always use a convenience method that includes `if T::NEEDS_TRACE { ... }`. To me this is not a very big deal, because:
  1) The previous technique only enforced this *in the case where there were multiple fields* due to moving the `Collection<'_>` object, and the fact that it doesn't work doesn't automatically hint to *why* things were designed that way.
  2) We can do whatever we want in the proc derive
  3) It usually not actually THAT bad if `Collect::trace` doesn't check `NEEDS_TRACE` on its field types. This only matters when the inner `Collect::trace` is not inline or doesn't itself early exit.

  Instead of `Collection::trace`, I added a `TraceExt` extension trait and a `TraceExt::trace` extension method which does what `Collection::trace` used to do. I would have liked to avoid an extension trait, but adding it as an inherent `Trace` method has problems as well, namely that it wouldn't work with `&mut dyn Trace` (which would be required for one day tracing pointee of `Gc<'gc, ()>`). I could have also added this convenience method on `Collect` in the form of an additional method that always checked `Self::NEEDS_TRACE`, but that is also somewhat confusing because both methods would have default implementations (and it wouldn't be immediately clear which one to implement). On the balance, this seemed the best option but I'm open to changing how this works.

* Should `Collect<'gc>` also imply `: 'gc`?
* Is `#[collect(gc_lifetime = 'gc)]` just overkill?
* The vtable `trace_value` function signature is `unsafe fn(GcBox, &mut Context)`, it *could* be `unsafe fn(GcBox, &mut dyn Trace<'gc>)` for future use, but I figured it was worth it to micro-optimize this?

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -202,7 +202,7 @@ fn collect_derive(s: synstructure::Structure) -> TokenStream {
                     // merge the spans. This is purely for diagnostic purposes, and has no effect
                     // on correctness
                     let bi = #bi;
-                    ::gc_arena::collect::TraceExt::trace(cc, bi);
+                    cc.trace(bi);
                 }
             )
         });
@@ -242,7 +242,7 @@ fn collect_derive(s: synstructure::Structure) -> TokenStream {
                     const NEEDS_TRACE: bool = #needs_trace_expr;
 
                     #[inline]
-                    fn trace<Trace: ::gc_arena::collect::Trace<#gc_lifetime> + ?Sized>(&self, cc: &mut Trace) {
+                    fn trace<Trace: ::gc_arena::collect::Trace<#gc_lifetime>>(&self, cc: &mut Trace) {
                         match *self { #trace_body }
                     }
                 }
@@ -253,7 +253,7 @@ fn collect_derive(s: synstructure::Structure) -> TokenStream {
                     const NEEDS_TRACE: bool = #needs_trace_expr;
 
                     #[inline]
-                    fn trace<Trace: ::gc_arena::collect::Trace<'gc> + ?Sized>(&self, cc: &mut Trace) {
+                    fn trace<Trace: ::gc_arena::collect::Trace<'gc>>(&self, cc: &mut Trace) {
                         match *self { #trace_body }
                     }
                 }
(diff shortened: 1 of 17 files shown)


kyren:
The `TraceExt` extension trait is gone in favor of an inherent method and *dropping* the `+ ?Sized` bound in `Collect::trace` everywhere (required because the inherent method must require `Self: Sized`).

This is not a big deal at all because `&mut dyn Trace` is ONLY going to be necessary for tracing through `Gc`, which can use an internal wrapper anyway (exactly how the current function works but not specialized to `&mut Context`).

PR #128, opened 2025-04-28
Add a few extra `Write` methods for `Box` and `[T]`

kyren:
Allows you to have a `Box<[Lock<T>]>` in a `Gc` and write to each element in the array.

In the future, when we allow directly allocating `dyn` types, the array write method will also allow you to have a `Gc<[Lock<T>]>` and write to a specific element.


kyren:
Superseded by #130

PR #129, opened 2025-06-13
Add `OnceLock` (a GC-aware `OnceCell`)

moulins:
The name clash with `std::sync::OnceLock` is a little unfortunate, but I can't think of a better name.

### DIFF
--- a/src/lock.rs
+++ b/src/lock.rs
@@ -1,7 +1,7 @@
 //! GC-aware interior mutability types.
 
 use core::{
-    cell::{BorrowError, BorrowMutError, Cell, Ref, RefCell, RefMut},
+    cell::{BorrowError, BorrowMutError, Cell, OnceCell, Ref, RefCell, RefMut},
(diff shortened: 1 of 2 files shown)


src/lock.rs
```diff
@@ -248,15 +252,7 @@ impl<T: fmt::Debug + ?Sized> fmt::Debug for RefLock<T> {
...
-                struct BorrowedPlaceholder;
-
-                impl fmt::Debug for BorrowedPlaceholder {
-                    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                        f.write_str("<borrowed>")
-                    }
-                }
-
-                fmt.field(&BorrowedPlaceholder)
+                fmt.field(&format_args!("<borrowed>"))
```
kyren, line 255: Somehow I've gone this long and never knew you could do this!

kyren:
This looks great, sorry this took so long for a review!

PR #130, opened 2025-06-13
Add `{Deref, Index}Write` traits, for deref. and indexing behind write barriers

moulins:
This is a more general version of #128.

It allows propagating write barriers through any type implementing the `DerefWrite` or `IndexWrite` traits. For example, modifying a value in a `Vec<Lock<T>>` can be written as such:
```rs
fn set_value<'gc, T>(mc: &Mutation<'gc>, gc: Gc<'gc, Vec<Lock<T>>>, v: T, idx: usize) {
    let write = Gc::write(mc, gc);
    write[idx].unlock().set(v);
}
```

### DIFF
--- a/src/allocator_api.rs
+++ b/src/allocator_api.rs
@@ -1,4 +1,4 @@
-use std::{alloc::Layout, marker::PhantomData, ptr::NonNull};
+use core::{alloc::Layout, marker::PhantomData, ops::Index, ptr::NonNull};
 
 use allocator_api2::{
     alloc::{AllocError, Allocator, Global},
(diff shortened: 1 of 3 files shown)


kyren:
Yeah this is a much better idea than doing it ad-hoc like #128.

I have thought about it as much as I can bear right now and can't think of any soundness complexities. The trait contracts for `DerefWrite` and `IndexWrite` make sense to me. I think it makes perfect sense to include `DerefMut` and `IndexMut` as part of the contracts even if we don't use that right now, the alternative of not including the Mut variants as part of the contract would be pretty weird?

Anyway this LGTM!

PR #136, opened 2026-01-27
Adds support for sized types with a dynamic layout

### LINKED ISSUE #106: impl From<Box<T>> for Gc<T> where T: ?Sized
It was mentioned in the README that this doesn't work yet, but I didn't see an issue -- it'd be nice to be able to go `Vec<T> -> Gc<[T]>`, but this seems like a  good first stepping-stone.

---

I gave the code a look to see how easy or hard this feels; I'm very new to this codebase, but my gut feeling is that this should be possible on nightly:

- Add an `unsafe fn Context::allocate_layout(Layout) -> NonNull<GcBoxInner<()>>`, which still constructs the header but leaves the contents uninitialized; safety conditions include something like "this must be initialized before returning from the enclosing Arena::mutate (or similar)"
- Expose this at `pub(crate)` in the Mutation
- Add a `From<Box<T>> where T: ?Sized`-like method that uses the `ptr_metadata` feature like so (untested):

```rust
impl<'gc, T: Collect + ?Sized + 'gc> Gc<'gc, T> {
    #[inline]
    pub fn from_box(mc: &Mutation<'gc>, boxed: Box<T>) -> Gc<'gc, T> {
        use core::{
            mem::ManuallyDrop,
            ptr::{addr_of_mut, copy_nonoverlapping, metadata, Pointee},
        };

        // Get the layout and metadata of the T.
        let layout = Layout::for_value::<T>(&boxed);
        let meta: <T as Pointee>::Metadata = metadata(&*boxed);

        // Create a Gc<'gc, ManuallyDrop<T>>. This has all sorts of safety obligations attached.
        // In particular, if a GC occurs while the boxed value is still uninitialized, UB will
        // occur during tracing.
        //
        // This really ought to be MaybeUninit instead, but that doesn't work with unsized types.
        let dst: NonNull<GcBoxInner<ManuallyDrop<T>>> = unsafe {
            let ptr: NonNull<GcBoxInner<()>> = mc.allocate_layout(layout);
            NonNull::from_raw_parts(ptr.cast(), meta)
        };

        // Make the Box be ManuallyDrop.
        let src: Box<ManuallyDrop<T>> = unsafe {
            let ptr = Box::into_raw(boxed);
            Box::from_raw(ptr as *mut ManuallyDrop<T>)
        };

        // Copy the value from src to dst, then free src. This won't drop the T, since we casted
        // the Box.
        unsafe {
            copy_nonoverlapping(
                &*src as *const ManuallyDrop<T> as *const u8,
                addr_of_mut!((*dst.as_ptr()).value) as *mut ManuallyDrop<T> as *mut u8,
                layout.size(),
            );
        }
        drop(src);

        // Cast away the ManuallyDrop.
        let ptr: NonNull<GcBoxInner<T>> = NonNull::from_raw_parts(dst.cast(), meta);

        // Return the nicely-wrapped-up pointer.
        Gc {
            ptr,
            _invariant: PhantomData,
        }
    }
}
```

I haven't looked deeply enough at `allocate_layout` to know whether unsizing the vtables will be a nightmare or not... The painful but probably sufficient approach, I suppose, would be to store the metadata in the `GcBoxInner`, along the lines of [`std::boxed::ThinBox`](https://doc.rust-lang.org/std/boxed/struct.ThinBox.html)'s `WithOpaqueHeader` type?

remexre:
This, with a healthy dose of unsafe, allows making thin pointers to
dynamically-sized values.

This changes the ABI of `Gc<T>` -- instead of being a pointer to the
header, it's a pointer to the value.

The header comes immediately behind the value in the allocation, without
padding. (The value pointer always has sufficient alignment for this.)
If a dynamic `Layout` is provided, the header will now have a Layout
before _it_ (again without padding, since the alignment matches).

```
      box
      ┌──────────────────────┐
      ┆  (padding, if needed)  ┆
      ├──────────────────────┤
      │ box layout, if dynamic │
      ├──────────────────────┤
    Gc│ header                 │
     ↳├──────────────────────┤
      │ value                  │
      └──────────────────────┘
```

---

I'm filing this PR as an RFC on these changes to the internals.

There's a few more changes I'd probably like to do before this is merged:

- [ ] Resolve #106 
- [ ] Add a better API around things like https://github.com/kyren/gc-arena/blob/383284c13352cb45c5a0febdd9378d28cbdebea8/tests/tests.rs#L1271-L1279 (taking an `init` callback that runs before we update `self.all` should be panic-safe; right now, this API pushes the panic safety obligation to the user)
- [ ] API for taking a wide-pointer `Box<T>` and storing the metadata in the allocation, along the same lines as [fat_type](https://docs.rs/fat_type/latest/fat_type/)? I don't actually need that for my use-case, but it's the more normal thing to want this API for.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ allocator-api2 = ["dep:allocator-api2", "hashbrown?/allocator-api2"]
 
 [dependencies]
 allocator-api2 = { version = "0.2", optional = true, default-features = false, features = ["alloc"] }
+contracts = { version = "0.6.7", default-features = false }
 gc-arena-derive = { path = "./derive", version = "0.5.3"}
 hashbrown = { version = "0.15.2", optional = true, default-features = false }
 tracing = { version = "0.1.37", optional = true, default-features = false }
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,16 +1,15 @@
-use alloc::{boxed::Box, vec::Vec};
+use alloc::{alloc::Layout, vec::Vec};
 use core::{
     cell::{Cell, UnsafeCell},
-    mem,
+    mem::{self, MaybeUninit},
     ops::{ControlFlow, Deref, DerefMut},
-    ptr::NonNull,
 };
 
 use crate::{
     Gc, GcWeak,
     collect::{Collect, Trace},
     metrics::Metrics,
-    types::{GcBox, GcBoxHeader, GcBoxInner, GcColor, Invariant},
+    types::{ErasedGcBox, GcBox, GcColor, Invariant},
 };
 
 /// Handle value given by arena callbacks during construction and mutation. Allows allocating new
@@ -40,19 +39,15 @@ impl<'gc> Mutation<'gc> {
     /// pointer(s) before collection is next triggered.
     #[inline]
     pub fn backward_barrier(&self, parent: Gc<'gc, ()>, child: Option<Gc<'gc, ()>>) {
-        self.context.backward_barrier(
-            unsafe { GcBox::erase(parent.ptr) },
-            child.map(|p| unsafe { GcBox::erase(p.ptr) }),
-        )
+        self.context
+            .backward_barrier(parent.ptr.erase(), child.map(|p| p.ptr.erase()))
     }
 
     /// A version of [`Mutation::backward_barrier`] that allows adopting a [`GcWeak`] child.
     #[inline]
     pub fn backward_barrier_weak(&self, parent: Gc<'gc, ()>, child: GcWeak<'gc, ()>) {
         self.context
-            .backward_barrier_weak(unsafe { GcBox::erase(parent.ptr) }, unsafe {
-                GcBox::erase(child.inner.ptr)
-            })
+            .backward_barrier_weak(parent.ptr.erase(), child.inner.ptr.erase())
     }
 
     /// IF we are in the marking phase AND the `parent` pointer (if given) is colored black, AND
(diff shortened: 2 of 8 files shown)


kyren:
I'm really sorry I just sort of left this PR for so long without giving you any feedback or input. 🙇‍♀️ I've only recently had a moment free to work on this crate in earnest again.

I looked over your PR briefly when I saw it the first time and also just now, and it is a solid idea. It is in fact such a solid idea that I just opened #150 to do something similar.

I'm commenting here because, if you're still interested and have time, could you take a look at #150 and see if it covers your use case?

remexre:
Great to see you back!

I'm kinda busy this week, but I'll try to give it a look over the weekend, if that's all right?

Just at a glance, though, my use-case actually needed more general support for unsized types than just a slice, e.g.

```rust
struct ThinBytesForSomeReason {
    size: usize,
    bytes: [u8],
}
```

(Actually, in particular, it mostly needs support for a fixed set of layouts that aren't known until runtime; it's the [compile-and-go system](https://en.wikipedia.org/wiki/Compile_and_go_system) in [my language](https://sylvan-lang.org/), which is in C right now.)

I could look at implementing this on top of your PR, or rebasing and cleaning up this PR, if it seems reasonable to try to land support for both together?

kyren:
> Just at a glance, though, my use-case actually needed more general support for unsized types than just a slice, e.g.

Did you take a look at `HeaderSlice`? That should be exactly this use case, and even if it's slightly annoying (you can't specify your own types) it should cover every use case of [fixed size header] + [trailing dyn slice].

For more advanced use cases, unfortunately supporting arbitrary `dyn` types is beyond the abilities of stable Rust atm BUT we're actually considering extending support for this *now* via a user accessible trait you can implement. I'm not sure yet whether this functionality will be in #150 or in a later PR, I'll have to experiment with it first.

kyren:
There is a new PR #152 that supports arbitrary unsized types, not just slices.

kyren:
Okay, FINALLY this should be fully subsumed by #152 and #155.

I believe `gc-arena` now supports dyn types to the greatest extend possible in stable Rust, with safe allocation of slices, strs, and `SliceWithHeader`. It should support allocating any user DST unsafely as long as you can get the metadata out of a fat pointer.

PR #139, opened 2026-05-28
Remove all notions of external allocation and allocator-api2 dependency

kyren:
The external allocation pacing was simply too complex to reason about.

This is a bit of an aggressive pruning, since now hashbrown collections that use allocator-api2 allocators do not automatically implement Collect, but this is probably a sufficiently niche case to not include.

What should replace this is custom user logic for external allocation and manual tuning of collection pacing. What would actually help here is if gc-arena *itself* allowed for custom allocators via allocator-api2.


kyren:
As a drive by change, updates some dependencies, the rust version used in CI, and updates the API tests.

PR #142, opened 2026-07-04
Implement `Collect` for `Pin<Ptr>` where `Ptr: Deref<Target: Collect>`

georgeclaghorn:
My immediate use case:

```rust
trait Foo<'gc>: DynCollect<'gc> {}

dyn_collect!(dyn Foo<'gc> + 'gc);

#[derive(Collect)]
#[collect(no_drop)]
struct Bar<'gc>(Pin<Box<dyn Foo<'gc> + 'gc>>);
```


georgeclaghorn:
Realized this is unsound.

PR #144, opened 2026-07-05
`dyn_collect!` macro examples: add missing `dyn` keyword

georgeclaghorn:
Using the macro as currently documented results in a compilation error:

```
error[E0782]: expected a type, found a trait
   --> runtime/src/execution/foo.rs:100:14
    |
100 | dyn_collect!(Foo<'gc>);
    |              ^^^^^^^^
    |
help: you can add the `dyn` keyword if you want a trait object
    |
100 | dyn_collect!(dyn Foo<'gc>);
    |              +++
```

I waffled on adding the keyword to the macro expansions instead of the examples (so the macro would work as documented). I went with changing the examples because the current behavior is [tested](https://github.com/kyren/gc-arena/blob/ae08ea7a243ddec4e2f0a717d6b798cd90e00ee1/tests/tests.rs#L1114).

### DIFF
--- a/src/collect.rs
+++ b/src/collect.rs
@@ -157,9 +157,9 @@ unsafe impl<'gc, T: Collect<'gc>> DynCollect<'gc> for T {
 /// This macro may be called in multiple ways:
 ///
 /// ```ignore
-/// dyn_collect!(MyTrait<'gc>);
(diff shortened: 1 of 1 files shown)


kyren:
Hey thank you!

I merged this, I just did it incorrectly because I can't use computers and then cherry-picked it into master.

PR #145, opened 2026-08-01
Use raw `Gc` count for collection metrics, ignore allocation size.

kyren:
Allocation size, especially without "external allocation" which was really hard to use and reason about, is almost meaningless.

A Rust struct with 10 `Copy` fields could easily contain fewer `Gc`s and cost less to allocate / free than a struct with one field which happens to be a `Vec`. Without some kind of allocator integration, this is unavoidable.

Raw `Gc` count is the most sensible metric we have until a *better* way of integrating with normal Rust allocation is found.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -3,11 +3,11 @@ use core::cell::Cell;
 
 /// Tuning parameters for a given garbage collected [`crate::Arena`].
 ///
-/// Any allocation that occurs during a collection cycle will incur "debt" that is exactly equal to
-/// the allocated bytes. This "debt" is paid off by running the collection algorithm some amount of
-/// time proportional to the debt. Exactly how much "debt" is paid off and in what proportion by the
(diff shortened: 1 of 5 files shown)


src/metrics.rs
```diff
@@ -49,10 +48,9 @@ use core::cell::Cell;
 ///    When this happens, `mark_factor * alloc_size` work is recorded.
 /// 2) Entries in the queue for tracing will eventually be traced by having their
 ///    [`crate::Collect::trace`] method called. At this time, `trace_factor * alloc_size` work is
```
moulins: You've missed a lot of `alloc_size` references in the main pacing documentation.

moulins (approved):
Excluding the above remark. LGTM.

src/metrics.rs
kyren: I sure did, I forgot to search for that one.

kyren: I did that and then also changed things referencing "size" to reference "count" instead

PR #148, opened 2026-08-02
New `ZstCache` type to perform ZST optimization for `Gc` allocation.

kyren:


zzz6519003 (commented):
why is it called ZstCache XD

kyren:
> why is it called ZstCache XD

It can be called something else? What do you have in mind?

PR #149, opened 2026-08-03
Change `Gc` and `GcWeak` to store pointers to `T` directly

kyren:
Instead of storing `NonNull<GcBoxInner<T>>` inside `Gc` and doing offsets during dereference, a `*const T` is stored. This has two primary benefits...

One, it makes dereferencing `Gc` as cheap as absolutely possible at the cost of making tracing *very slightly* more expensive (equal to the decreased expense of `Gc` dereference). Since `Gc` dereferencing is likely to be vastly more common, this is a win.

Two, it makes it possible to safely cast from `Gc<T>` to `Gc<U>` with the only requirement being that it must be valid to cast `*const T` to `*const U` and dereference it. Before, this is not sufficient, as both types must also have the same alignment so that `GcBoxInner<T>` and `GcBoxInner<U>` are guaranteed to have the inner value at the same location.

The cost of allocation should never be worse than it was before. The layout of GC allocations is simply changed so that the header is always *just* before the value, and any required padding is *before* the header rather than after. This way, finding the header is always subtracting the fixed header size from the value pointer, and this is always correct no matter the held value type.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,77 +1,148 @@
+use alloc::alloc;
 use core::alloc::Layout;
 use core::cell::Cell;
 use core::marker::PhantomData;
 use core::ptr::NonNull;
-use core::{mem, ptr};
+use core::{fmt, mem, ptr};
 
 use crate::{collect::Collect, context::Context};
 
 /// A thin-pointer-sized box containing a type-erased GC object.
-/// Stores the metadata required by the GC algorithm inline (see `GcBoxInner`
-/// for its typed counterpart).
+///
+/// Stores the metadata required by the GC algorithm in the same allocation *before* the stored
+/// pointer.
+pub(crate) struct GcBox<T: ?Sized = ()>(NonNull<T>);
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub(crate) struct GcBox(NonNull<GcBoxInner<()>>);
-
-impl GcBox {
-    /// Erases a pointer to a typed GC object.
-    ///
-    /// **SAFETY:** The pointer must point to a valid `GcBoxInner` allocated
-    /// in a `Box`.
-    #[inline(always)]
-    pub(crate) unsafe fn erase<T: ?Sized>(ptr: NonNull<GcBoxInner<T>>) -> Self {
-        // This cast is sound because `GcBoxInner` is `repr(C)`.
-        let erased = ptr.as_ptr() as *mut GcBoxInner<()>;
-        unsafe { Self(NonNull::new_unchecked(erased)) }
+impl<T: ?Sized> fmt::Debug for GcBox<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Pointer::fmt(&self.as_ptr(), f)
     }
+}
 
-    /// Gets a pointer to the value stored inside this box.
-    /// `T` must be the same type that was used with `erase`, so that
-    /// we can correctly compute the field offset.
+impl<T: ?Sized> Copy for GcBox<T> {}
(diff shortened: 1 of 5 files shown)


src/types.rs
```diff
@@ -1,77 +1,239 @@
...
 
-    /// Gets a pointer to the value stored inside this box.
-    /// `T` must be the same type that was used with `erase`, so that
-    /// we can correctly compute the field offset.
+impl<'gc, T: Collect<'gc>> GcBox<T> {
+    /// Allocate a new `GcBox` with a default header and holding the given value.
     #[inline(always)]
-    fn unerased_value<T>(&self) -> *mut T {
+    pub(crate) fn alloc(value: T) -> Self {
         unsafe {
```
moulins: It would be slightly more performant to take a `GcBoxHeader` parameter here, so that it doesn't need to be adjusted afterwards in `Context::allocate`.

src/types.rs
```diff
@@ -1,77 +1,239 @@
...
-            let ptr = self.0.as_ptr() as *mut GcBoxInner<T>;
-            // Don't create a reference, to keep the full provenance.
-            // Also, this gives us interior mutability "for free".
-            ptr::addr_of_mut!((*ptr).value) as *mut T
+            let gc_header = GcBoxHeader::new::<T>();
+            let gc_layout = gc_header.vtable().gc_layout;
+
+            let bytes = alloc::alloc(gc_layout.alloc_layout);
+            let bytes = NonNull::new(bytes).expect("allocation error");
```
moulins: This should be `NonNull::new(bytes).unwrap_or_else(|| alloc::handler_alloc_error(gc_layout)` instead.

src/types.rs
```diff
@@ -184,12 +356,12 @@ impl GcBoxHeader {
...
+    trace_value: unsafe fn(*const (), &mut Context),
+    /// Drops the value at the given pointer.
+    drop_value: unsafe fn(*mut ()),
     /// The layout of the `GcBox` the GC'd value is stored in.
-    box_layout: Layout,
-    /// Drops the value stored in the given `GcBox` (without deallocating the box).
-    drop_value: unsafe fn(GcBox),
-    /// Traces the value stored in the given `GcBox`.
-    trace_value: unsafe fn(GcBox, &mut Context),
+    gc_layout: GcLayout,
```
moulins: I don't think storing the `value_offset` in the vtable is necessary? we can recompute it from the "bare" `Layout` at deallocation time. (by changing `GcBoxHeader::new` to take a `Layout` parameter instead of a generic)

src/types.rs
kyren: This is fixed, thanks for teaching me about that API!

src/types.rs
kyren: You meant `GcLayout::new` to take a Layout parameter, right? If so that makes sense and is fixed now.

src/types.rs
kyren: The only issue is that it makes `alloc` unsafe because you have to pass in the correct GcBoxHeader for the correct `T`.

I understand everything having to do with this is unsafe but it would be nice not to have extra ways to fail. Plus, the whole thing is *definitely* going to be inlined so I doubt it'd matter?

kyren:
In two separate commits I did the following drive-by changes...

1) Updated the doc comments to `Gc` to state that we are storing a pointer to `T` directly, and while I was at it I added doc comments for `GcWeak` at *all*.

2) I did some renames of internal types whose names have bothered me for a while. Mainly it changes `GcBox` -> `GcPtr` since it's 1) not a `Box`, and 2) is now used everywhere instead of `NonNull<GcBoxInner<T>>` so it feels worse.

PR #150, opened 2026-08-04
Add the ability to allocate slices directly in a Gc.

kyren:
Adds a type called `HeaderSlice` which is the shape that all allocated
slices must fit. It allows for an arbitrary (sized) header type before
the slice.

Allocated slices take up an extra machine word in memory because they
store the slice length as a prefix to the normal GC header.


src/context.rs
```diff
@@ -355,20 +368,62 @@ impl Context {
...
+
+            impl<H, E> Drop for DropGuard<H, E> {
+                fn drop(&mut self) {
+                    unsafe {
+                        let uninit_mut = self.uninit_ptr.as_mut();
+
+                        uninit_mut.header.assume_init_drop();
+                        for i in 0..self.init_length {
+                            uninit_mut.slice[i].assume_init_drop();
+                        }
```
moulins: This can be simplified to:
```rs
let ptr = HeaderSlice::to_thin_ptr(self.uninit_ptr.as_ptr) as *const H;
let ptr = HeaderSlice::<H, E>::from_thin_ptr(ptr, self.init_length);
core::ptr::drop_in_place(ptr);
```

src/header_slice.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        let array_layout = match Layout::array::<E>(len) {
+            Ok(l) => l,
+            Err(e) => return Err(e),
+        };
+
+        match header_layout.extend(array_layout) {
+            Ok((layout, _)) => Ok(layout.pad_to_align()),
+            Err(e) => Err(e),
+        }
```
moulins, line 66: Missing `?` in const is sadness :c

src/gc_ptr.rs
```diff
@@ -0,0 +1,586 @@
...
+    ///
+    /// Panics if a `HeaderSlice<H, E>` of length `len` does not have a valid layout.
+    #[inline]
+    pub(crate) fn alloc_slice(len: usize) -> Self {
+        unsafe {
+            let slice_layout =
+                HeaderSlice::<H, E>::layout(len).expect("no layout for slice of requested length");
+
+            let ph_layout = PrefixHeaderLayout::new(GC_LEN_THEN_HEADER_LAYOUT, slice_layout)
+                .expect("no layout for GC allocation");
```
moulins: I would merge the two panics here; the exact point where the layout overflows doesn't really matter.

src/gc_ptr.rs
```diff
@@ -0,0 +1,586 @@
...
+/// Layout of a block of memory containing a length and `GcHeader`.
+///
+/// The length will be stored at the *beginning* of this block and the header at the *end* (which
+/// would only matter in the odd case where `GcHeader`'s alignment is less than that of `usize`).
+const GC_LEN_THEN_HEADER_LAYOUT: Layout =
+    if let Ok((layout, _)) = Layout::new::<usize>().extend(GC_HEADER_LAYOUT) {
+        layout
+    } else {
+        unreachable!();
+    };
```
moulins, line 315: `.ok().unwrap()` could be used to avoid manual matching (`Option::unwrap` is const, unlike `Result::unwrap`)

moulins (commented):
I have to say I'm having trouble wrapping my head around the logic in `PrefixHeaderLayout`... I'll think more about it and make some diagrams '^^

src/header_slice.rs
kyren: so true 😔

src/gc_ptr.rs
kyren: I did not think of that 🫨 but it makes perfect sense.

kyren: `Result::ok()` is not yet stable as a const fn 😩

kyren:
@moulins I implemented all of your suggestions (except the one with the unstable const fn).

I also went ahead and changed things so that the `Context` allocation functions returned *uninitialized* pointers and moved the initialization into gc.rs. This also enabled me to implement `Copy` versions of the slice constructors via `ptr::copy_nonoverlappying`, and should make a "builder API" much more feasible.

kyren:
> I have to say I'm having trouble wrapping my head around the logic in PrefixHeaderLayout... I'll think more about it and make some diagrams '^^

I keep feeling like the logic could be simpler but every time I think that I realize my idea for simplifying it is wrong. If you can come up with a simpler way to express it that would be welcome.

kyren:
I think there was a very subtle bug in `PrefixHeaderLayout` that would only trigger if the alignment of the header was less than the alignment of a `usize` (even though this is currently impossible). I've added documentation and asserts for the precondition that the entire header layout must have its size as a multiple of its alignment.

moulins:
> I think there was a very subtle bug in PrefixHeaderLayout that would only trigger if the alignment of the header was less than the alignment of a usize (even though this is currently impossible). I've added documentation and asserts for the precondition that the entire header layout must have its size as a multiple of its alignment.

Actually, after a lot of pondering, I believe `PrefixHeaderLayout::new` can be simplified to this, without requiring any new preconditions (unless I missed something obvious...):
```rust
const fn new(header_layout: Layout, value_layout: Layout) -> Option<Self> {
  if let Ok((alloc_layout, value_offset)) = header_layout.extend(value_layout) {
    Self { alloc_layout, value_offset }
  } else {
    None
  }
}
```
- even though the header won't be placed at offset 0, `Layout::extend` still gives us the correct offset for the value;
- No need to call `pad_to_align()` on `alloc_layout`, as this isn't the layout of an actual Rust struct, and such doesn't need to respect the "size multiple of alignment" rule.

kyren:
> Actually, after a lot of pondering, I believe `PrefixHeaderLayout::new` can be simplified to this, without requiring any new preconditions (unless I missed something obvious...):
> 
> ```rust
> const fn new(header_layout: Layout, value_layout: Layout) -> Option<Self> {
>   if let Ok((alloc_layout, value_offset)) = header_layout.extend(value_layout) {
>     Self { alloc_layout, value_offset }
>   } else {
>     None
>   }
> }
> ```
> 
>     * even though the header won't be placed at offset 0, `Layout::extend` still gives us the correct offset for the value;

I think you're right except I believe you still need the precondition that the header's size is a multiple of its alignment.

Consider a header of size 7, alignment 4 and a value of size 8 alignment 2.

If the header's size is always a multiple of its alignment, then I think you're right? It works because greater alignments are always multiples of lesser alignments, so if header's alignment is greater then value will be placed adjacent to the header and everything is fine. If value's alignment is greater than subtracting the header's size always yields the header's alignment.

The header is also not necessarily a plain Rust struct so wherever that is calculated it needs to include a `pad_to_align()` for everything to work out.

The old implementation was doing this:

```rust
        let alloc_align = max(header_align, value_align);
        let value_offset = max(header_size, value_align);
        let Some(alloc_size) = value_offset.checked_add(value_size) else {
            return None;
        };
        let Ok(alloc_layout) = Layout::from_size_align(alloc_size, alloc_align) else {
            return None;
        };
```

Which... I think is just basically the contents of `Layout::extend` but instead of `size_rounded_up_to_custom_alignment` it was doing `max(header_size, value_align)` which is obviously wrong 😩.

>     * No need to call `pad_to_align()` on `alloc_layout`, as this isn't the layout of an actual Rust struct, and such doesn't need to respect the "size multiple of alignment" rule.

Yeah that's true, it wasn't doing that before either.

kyren:
> I think there was a very subtle bug in PrefixHeaderLayout that would only trigger if the alignment of the header was less than the alignment of a usize (even though this is currently impossible). I've added documentation and asserts for the precondition that the entire header layout must have its size as a multiple of its alignment.

Just so everything makes sense, the reason I said that was because in this PR that's the only way you'd get a header size that *wasn't* a multiple of its alignment since it was computed like `Layout::new::<usize>().extend(Layout::new::<RestOfHeader>())`, when it should have been `Layout::new::<usize>().extend(Layout::new::<RestOfHeader>()).pad_to_align()`. The original change I made added this and added asserts that it was true.

Of course I was missing the really obvious bug in the layout calculation but *whatever*.

PR #152, opened 2026-08-06
Support arbitrary unsized `Gc` types and type-level metadata.

kyren:
A version of #150 that goes much further to support arbitrary unsized types and other vtable tricks.

Right now this is only safely usable with `SliceWithheader`, but in the future it should be very easy to extend this to support `Pointee` type metadata if / when that is stabilized.


src/context.rs
```diff
@@ -367,33 +358,18 @@ impl Context {
...
+    /// metrics.
     #[inline]
-    fn allocate<'gc, T: Collect<'gc>>(&self) -> GcPtr<T> {
-        let gc_ptr = GcPtr::<T>::alloc();
-        gc_ptr.header().set_needs_trace(T::NEEDS_TRACE);
-        self.link_ptr(gc_ptr);
-        gc_ptr
-    }
+    fn link<'gc, T: ?Sized>(&self, gc_ptr: GcPtr<T>) {
+        gc_ptr.header().set_next(self.all.get());
```
moulins: This doesn't need to be generic, it can take an erased `GcPtr`.

src/dynamic_roots.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
 //
 // SAFETY: This allows us to convert `Gc<'gc>` pointers to `Gc<'static>` and back, and this is VERY
 // sketchy. We know it is safe because:
-//   1) The `DynamicRootSet` must be created inside an arena and is branded with an invariant `'gc`
+//   1) The `Dync`
```
moulins: Oh noes! what horrible fate befell this poor line ? :((

src/gc.rs
```diff
@@ -304,75 +246,150 @@ impl<'gc, T: 'static> Gc<'gc, T> {
...
+    /// Similarly to `Rc::ptr_eq` and `Arc::ptr_eq`, this function ignores the metadata of `dyn`
+    /// pointers.
+    #[inline]
+    pub fn ptr_eq(this: Self, other: Self) -> bool {
+        this.ptr.addr_eq(other.ptr)
+    }
+
+    /// Cast a `Gc` to the unit type.
+    ///
+    /// This converts the `Gc` to point to a `()`, and converts changes the kind to
```
moulins: typo: `converts changes`

src/gc.rs
```diff
@@ -304,75 +246,150 @@ impl<'gc, T: 'static> Gc<'gc, T> {
...
+    /// [`DefaultGcKind`].
+    ///
+    /// This is always safe to do as it is always save to dereference a `*const ()` which comes from
```
moulins: typo: `save` -> `safe`

src/slice.rs
```diff
@@ -0,0 +1,353 @@
...
+    ) -> GcSliceWithHeader<'gc, H, E, M> {
+        unsafe {
+            ptr::copy_nonoverlapping(
+                elements.as_ptr(),
+                self.slice_ptr() as *mut E,
+                elements.len(),
+            );
+            self.assume_init(mc)
+        }
+    }
```
moulins, line 244: This is unsound: we need to check that `elements` has the correct length (and panic if not, or something).

moulins (commented):
The safety contracts on `PtrMeta` and `AllocMeta` are missing, but they can be added later so... 🤷

src/context.rs
kyren: oh of course, don't know why it's like that

src/dynamic_roots.rs
kyren: oops 🤭

src/slice.rs
kyren: lost in refactoring 😩 thank you for catching this

